PR #68, opened 2017-08-22
Suppress warnings in curve25519.tweetnacl.c

huitema:
The changes ought to be pretty small. There does not appear to be any effect on test. But look at the change in the function sel2119. It was previously:

> static void sel25519(gf p, gf q, int b)
> {
>    int64_t tmp, mask = ~(b-1);

That is, the argument "b" was declared as int and used as int64. This cannot be right, since the definition of int can vary with the machine architecture. I changed that to:

> static void sel25519(gf p, gf q, int64_t b)
> {
>    int64_t tmp, mask = ~(b-1);

This suppresses the warning. I hope that it respects the original intent.

### DIFF
--- a/deps/cifra/src/curve25519.tweetnacl.c
+++ b/deps/cifra/src/curve25519.tweetnacl.c
@@ -57,7 +57,7 @@ static void car25519(gf o)
   }
 }
 
-static void sel25519(gf p, gf q, int b)
+static void sel25519(gf p, gf q, int64_t b)
 {
   int64_t tmp, mask = ~(b-1);
   size_t i;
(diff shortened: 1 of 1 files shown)


kazuho:
Merging the PR since I believe that this is the one of the right ways to suppress the warning.

> That is, the argument "b" was declared as int and used as int64. This cannot be right, since the definition of int can vary with the machine architecture.

FWIW, I believe that the intent of `mask = ~(b-1)` is to convert a boolean value `b` into either a bitmask that has all the bits set (if `b` is one), or one that has all the bits cleared (if `b` is zero). Hence the original code is _right_. However, it is still a good idea to adjust the code to suppress warnings generated by a compiler.

huitema:
Yes, the code was obviously correct, and specifying the argument as ont was not wrong. The other way to suppress the warning is to explicitly cast the 3rd parameter to (int) in the function calls. That might result in slightly faster code in 32bit processors, but I assume that compiler optimisations would take care of that.

-- Christian Huitema

PR #88, opened 2017-11-28
Update Visual Studio Solution to support both 32 bit (WIN32) and 64 bits (X64)

huitema:
Most of the changes are in the Visual Studio project files. There are however also a few changes in the code file, of two natures:

1) Replace "#ifdef WIN32" by "#ifdef _WINDOWS". The key WIN32 specifically refers to the x86 architecture, so is not adequate if we want to compile both 32 bits (x86) and 64 bits (x64).

2) Fix a small number of compile warning that happens when compiling in 64 bit mode. Mostly, these are related to using "size_t", which on Windows is defined as "uint32_t" for x86, but "uint64_t" on x64.


lib/pembase64.c
```diff
@@ -247,7 +247,7 @@ static int ptls_compare_separator_line(const char *line, const char *begin_or_en
     if (ret == 0) {
         size_t begin_or_end_length = strlen(begin_or_end);
         ret = strncmp(line + text_index, begin_or_end, begin_or_end_length);
-        text_index += begin_or_end_length;
+        text_index += (int) begin_or_end_length;
```
kazuho: Would it make sense to change type of `text_index` to `size_t` instead?

huitema:
The travis check fails because I need to update the submodule deps/picotest, and I probably did not do it correctly. Any suggestion?

huitema:
Kazuho: I have pushed a separate PR for h2o/picotest. Can you take a look please? I think that once that PR is merged, I will be able to push this one.

kazuho:
@huitema I've merged https://github.com/h2o/picotest/pull/2.

Am I correct in believing that the issue has been resolved already? It seems that you have already merged this PR and the tests have passed on Travis CI. Please let me know if you need any help.

huitema:
Yes. I had to roll back the commit that  tried to update deeps/picotest, and was failing. No moved that commit directly to the picotest depot, and the reminder worked fine.

Thanks for merging the PR.

-- Christian Huitema

Issue #85, opened 2017-11-25
Memleak related to ctx.certificates


Lekensteyn:
ctx.certificates (in `ptls_context_t`) is malloc'ed memory and its contents (`dst->base` via `i2d_X509` in `t/util.h`) is not freed. This results in several memleak report from ASAN:
```
$ ./cli -4 localhost 4433 -c server.pem -h
Usage: ./cli [options] host port
...
==14645==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 328 byte(s) in 1 object(s) allocated from:
    #0 0x7fa446e62ae9 in __interceptor_malloc /build/gcc-multilib/src/gcc/libsanitizer/asan/asan_malloc_linux.cc:62
    #1 0x7fa4467f4639 in CRYPTO_zalloc (/usr/lib/libcrypto.so.1.1+0x15a639)
    #2 0x7fa446728d07  (/usr/lib/libcrypto.so.1.1+0x8ed07)
    #3 0x7fa446726772  (/usr/lib/libcrypto.so.1.1+0x8c772)
    #4 0x7fa4467272ba in ASN1_item_ex_d2i (/usr/lib/libcrypto.so.1.1+0x8d2ba)
    #5 0x7fa44672733b in ASN1_item_d2i (/usr/lib/libcrypto.so.1.1+0x8d33b)
    #6 0x7fa446807a20 in PEM_ASN1_read_bio (/usr/lib/libcrypto.so.1.1+0x16da20)
    #7 0x7fa446805c48 in PEM_ASN1_read (/usr/lib/libcrypto.so.1.1+0x16bc48)
    #8 0x5602d2974564 in load_certificate_chain /home/peter/projects/picotls/t/util.h:51
    #9 0x5602d2974564 in main /home/peter/projects/picotls/t/cli.c:322
    #10 0x7fa4453dff69 in __libc_start_main (/usr/lib/libc.so.6+0x20f69)
    #11 0x5602d2975489 in _start (/tmp/pico/cli+0x52489)
...
```

huitema:
Do the leaks persist if instead of calling the openssl library you use ptls_load_certificates()?

Lekensteyn:
@huitema Leaks are gone when using the built-in routines, done in #99

PR #99, opened 2018-01-10
Fix memleak and potential crash in cli

Lekensteyn:
Fix for memleak issue reported in #85 as well as a related issue I found while testing with a certificate which contains multiple entries.

### DIFF
--- a/lib/pembase64.c
+++ b/lib/pembase64.c
@@ -299,7 +299,7 @@ static int ptls_get_pem_object(FILE *F, const char *label, ptls_buffer_t *buf)
     return ret;
 }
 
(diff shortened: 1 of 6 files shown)


lib/pembase64.c
```diff
@@ -328,8 +328,8 @@ int ptls_load_pem_objects(char const *pem_fname, const char *label, ptls_iovec_t
 
             if (ret == 0) {
                 if (buf.off > 0 && buf.is_allocated) {
-                    list[count]->base = buf.base;
-                    list[count]->len = buf.off;
+                    list[count].base = buf.base;
+                    list[count].len = buf.off;
                     count++;
```
huitema: Damn. My bug. You are correct, the ** was one * too many. Thanks for fixing that.

PR #170, opened 2018-08-03
Fix search path in x64/Release VS project

huitema:


huitema:
Merging this, as it only affects the VS project.

Issue #168, opened 2018-08-03
Unit tests failing on Windows 64 bits


huitema:
When running Picoquic on Windows, @lzq8272587 noticed that the Windows port of the unit tests failed. Here is his original message:

---------------------------
Hi huitema:

    I try to compile and run the code on Visual Studio 2017, but I get the same error stack as VS2015.

Something must be wrong in calling read_entropy.

    I also tried to run testopenssl, and a similar issue happens.

First, this is the output in my console:

# Subtest: key-exchange
ok 1 - e:\picotls\picotls-master\t\picotls.c 1067
ok 2 - e:\picotls\picotls-master\t\picotls.c 1071
ok 3 - e:\picotls\picotls-master\t\picotls.c 1073
ok 4 - e:\picotls\picotls-master\t\picotls.c 1075
ok 5 - e:\picotls\picotls-master\t\picotls.c 1076
ok 6 - e:\picotls\picotls-master\t\picotls.c 1077
ok 7 - e:\picotls\picotls-master\t\picotls.c 1085
ok 8 - e:\picotls\picotls-master\t\picotls.c 1087
ok 9 - e:\picotls\picotls-master\t\picotls.c 1088
ok 10 - e:\picotls\picotls-master\t\picotls.c 1067
ok 11 - e:\picotls\picotls-master\t\picotls.c 1071
ok 12 - e:\picotls\picotls-master\t\picotls.c 1073
ok 13 - e:\picotls\picotls-master\t\picotls.c 1075
ok 14 - e:\picotls\picotls-master\t\picotls.c 1076
ok 15 - e:\picotls\picotls-master\t\picotls.c 1077
ok 16 - e:\picotls\picotls-master\t\picotls.c 1085
ok 17 - e:\picotls\picotls-master\t\picotls.c 1087
ok 18 - e:\picotls\picotls-master\t\picotls.c 1088
ok 19 - e:\picotls\picotls-master\t\picotls.c 1067
ok 20 - e:\picotls\picotls-master\t\picotls.c 1071
ok 21 - e:\picotls\picotls-master\t\picotls.c 1073
ok 22 - e:\picotls\picotls-master\t\picotls.c 1075
ok 23 - e:\picotls\picotls-master\t\picotls.c 1076
ok 24 - e:\picotls\picotls-master\t\picotls.c 1077
ok 25 - e:\picotls\picotls-master\t\picotls.c 1085
ok 26 - e:\picotls\picotls-master\t\picotls.c 1087
[...]

PR #174, opened 2018-08-07
Replace wincrypt by bcrypt on modern Windows systems

huitema:
This is a fix for issue #168.

The unit tests were failing for @lzq8272587 because of his windows system configuration. The failure happened when instantiating the "wincrypt" API before using CryptGenRandom in order to "read entropy" on Windows systems. The old "wincrypt" API requires opening a "security container" through the "CryptAcquireContext" API, and on LZQ system the access control configuration prevented user processes from doing that.

The fix is to use the more modern "bcrypt" API. The call to "BCryptGenRandom" does not require a security context, and thus is not subject to access control. 

The downside is that the "bcrypt" API is not available for Windows XP. We leave the old "wincrypt" code available under the "_WINDOWS_XP" compile flag, should someone really care about compiling Picotls on Windows XP. 

Using the "bcrypt" API requires linking with "bcrypt.lib".

### DIFF
--- a/lib/cifra.c
+++ b/lib/cifra.c
@@ -44,7 +44,11 @@
 #include "picotls/minicrypto.h"
 
 #ifdef _WINDOWS
+#ifdef _WINDOWS_XP
+ /* The modern BCrypt API is only available on Windows Vista and later versions.
+  * If compiling on Windows XP, we need to use the olded "wincrypt" API */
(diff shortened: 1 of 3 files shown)


huitema:
@kazuho: I am merging this PR as it only affects the Windows code and solves a blocking issue for @lzq8272587.

PR #206, opened 2019-01-28
parser: fail early on invalid client hello

toru:
## Objective

Fail early in `handle_input`  if the parsed client hello record is invalid.

## How

Instead of proceeding on `PTLS_ERROR_IN_PROGRESS`, check the parsed header to determine if it's necessary to continue with the given input. If invalid, fail early by returning an alert.

Reported by David Wong.


toru:
Closing in favor of revisiting with a more general approach.

PR #237, opened 2019-04-01
Fix bench bug

huitema:
Series of simple fixes for benchmark issues when running on Ubuntu.


huitema:
Merging without review, since this are minor bug fixes in a test tool.

PR #253, opened 2019-07-08
Add ability to override read_entropy function

larseggert:


larseggert:
This has become redundant with #254, since by not linking cifra.c, I can simply substitute my own version of `ptls_minicrypto_random_bytes`

PR #277, opened 2019-10-24
Fix bug in Visual Studio 32 bit config for bcrypt project

huitema:


huitema:
Merging, as this is a minor bug fix only affecting Visual Studio 32 bit builds.

PR #293, opened 2020-03-03
Add Cmake install directives

sknat:
Add position independent flag & install directives for use in quicly


epsilon-0:
This is a nice functionality to have as I'm adding this for Gentoo (in guru overlay). Is there anything that I can help to merge this PR?

sknat:
Thanks @epsilon-0 this is an old patch that has gone stale. But it seems to still be mergeable.
Best would probably to ask @kazuho if this seems acceptable.

PR #323, opened 2020-09-12
Confidentiality and integrity limits for AEAD

huitema:
Add two constants to the definition of AEAD algorithms: the confidentiality limit and the integrity limit. This enable implementations of QUIC to find these limits as part of the AEAD objects, and is extensible to new AEAD types if implementations add them.


kazuho (commented):
Thank you for the PR. Merged.

PR #352, opened 2021-03-28
Adding win-compile action

huitema:
Or at least trying to...

### DIFF
--- a/.github/workflows/win_compile.yml
+++ b/.github/workflows/win_compile.yml
@@ -13,27 +13,12 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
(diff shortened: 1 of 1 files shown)


huitema:
@kazuho This "github action" appears to be a rathole. To build picotls I need openssl, but there is no clean way to install the openssl libraries on windows. So I need to clone openssl and build it. I got the cloning part done, and also the configuration, but to compile openssl on windows I need to install the `nmake` tool and I just can't find a way to do that. The openSSL documentation says "find the VS developer command line tool" but there does not appear to be a good way to do that in a script. Besides, building OpenSSL from scratch for every run seems very costly.

I have scripts working with `appveyor` for Picoquic. I will port them in another PR.

PR #375, opened 2021-12-13
More compatible with Windows

BernardoGomesNegri:
This changes a few bits of the code to make the library compatible with Windows. It changes some included header files to point to the Windows equivalents, as well as adding some more definitions that are not avaliable on Windows that should act as a drop-in replacement for their POSIX equivalents.

### DIFF
--- a/lib/picotls.c
+++ b/lib/picotls.c
@@ -25,7 +25,9 @@
 #include <stdlib.h>
 #include <string.h>
 #ifdef _WINDOWS
-#include "wincompat.h"
+#include <ws2tcpip.h>
+#include <winsock.h>
+#include <winsock2.h>
 #else
 #include <arpa/inet.h>
 #include <sys/time.h>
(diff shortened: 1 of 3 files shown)


BernardoGomesNegri:
CI failed, but it does work with mingw.

lib/picotls.c
```diff
@@ -112,6 +114,36 @@
...
+    SystemTimeToFileTime( &system_time, &file_time );
+    time =  ((uint64_t)file_time.dwLowDateTime )      ;
+    time += ((uint64_t)file_time.dwHighDateTime) << 32;
+
+    tp->tv_sec  = (long) ((time - EPOCH) / 10000000L);
+    tp->tv_usec = (long) (system_time.wMilliseconds * 1000);
+    return 0;
+}
+#endif
+
```
huitema, line 146: gettimeofday is also used in ptlsbench. Are you going to leave that code use the old "wincompat.h", or do you want to export the signature and the defintion of timezone in another header?

huitema, line 146: The replacement code proposed here is already defined in "wintimeofday.c", and the code has been used for the Visual Studio builds for a long time. We should not have two versions of that code in the same project. Also, the existing code uses "GetSystemTimePreciseAsFileTime", which achieves higher precision.

huitema (changes requested):
What problem are you trying to solve? The xisting visuall studio code works well and has been in use for a long time. You may want to streamline it, but if you do that you should ensure that there is no regression in Visual Studio, and also no duplication of code in the resulting set.

For example: reconcile the proposed inclusion of gettimeofday with the code existing in "wintimeofday.c", which uses a more precise defintion of time than the proposed replacement. Also, track all usages of "wincompat.h", and either replace them all or replace none...

PR #396, opened 2022-07-09
Check new openssl api

huitema:
Add flag in header file openssl.h, so applications can check which version of the API to use.


huitema:
Also fix the memory free in Windows. It seems that something changed, and that the certificates are now loaded in memory aligned buffers, requiring special treatment.

Issue #412, opened 2022-09-02
unable to build with openssl1.1 on windows


hd51515:
[ptlsvserrorwithopenssl111.txt](https://github.com/h2o/picotls/files/9478397/ptlsvserrorwithopenssl111.txt)

huitema:
@hd51515 I think that PR #413 solves your issues. Can you verify please?

huitema:
@hd51515 I had to delete PR #413 because of GitHub issues. Please try PR #415 instead.

huitema:
Close as PR #415 is merged.

PR #413, opened 2022-09-03
Openssl3 vs17

huitema:
Addresses issue #412


huitema:
@kazuho -- it seems that I picked the unrelated changes for removing the compile warnings. I could separate them if you want.

PR #431, opened 2022-11-02
[ptlslog] add a configuration, `ptls_log.max_str_len`, to limit the size of unsafe and binary values

gfx:
In ptls logs, unsafe values and hex-dumped binary values might be large and thus might cause dynamic allocations in `PTLS_LOG__DO_LOG`. This feature mitigates such situations by limiting the size of unsafe and binary values. I don't think we need to apply this limitation to "safe" value, because they all should be expected, and thus we don't want to truncate them.

### DIFF
--- a/include/picotls.h
+++ b/include/picotls.h
@@ -1307,6 +1307,12 @@ uint64_t ptls_decode_quicint(const uint8_t **src, const uint8_t *end);
 typedef struct st_ptls_log_t {
     unsigned is_active : 1;
     unsigned include_appdata : 1;
+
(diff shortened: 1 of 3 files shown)


gfx:
Discussed with @kazuho elsewhere and we've agreed that we don't need to make it configurable. It is the client (i.e. use case) that should manage the overhead of ptlslog, not the process that uses ptlslog.

Instead, we'll implement knobs to enable/disable per probe type when we control the overhead of ptlslog. 

Closing.

PR #462, opened 2023-01-31
fix missing client_ciphers when calling the client hello cb

robguima:
Sorry, but  it looks like missed adding the client_ciphers for the constraints case in https://github.com/h2o/picotls/pull/227

So, when `client_hello_constraints` is called those aren't passed along to `call_on_client_hello_cb`.


robguima:
`ptls_decode16` is also `ntoh16` which is not super helpful, but very minor (depends on how the consumer wants it).

Issue #317, opened 2020-07-14
Installing picotls?


huitema:
Is there a way to do something like "make install" and install the picotls libraries in the system so they can easily be used by other projects?

PR #475, opened 2023-04-03
Add support for `make install`.

ioquatix:
I don't know a lot about CMake but I tested this locally and it seems to work.


ioquatix:
Fixes https://github.com/h2o/picotls/issues/317

PR #528, opened 2024-05-16
Complete implementation of MbedTLS as backend

huitema:


barracuda156:
@huitema How is this going?

huitema:
I think this PR is ready. It allows using MbedTLS as a self-sufficient back end, including for functions like certificate verification. I would like review of the code that derives the server's public key from the list of certificates. In all the tests, the size of the list is 1, so the assumption that the first certificate is good works. But if the list contains more certificates, we probably have some extra work to do.

huitema:
@kazuho in the test assets, do we have example of certificate chains containing more than 1 certificate?

doublex:
@kazuho 
Are there any reasons against this pull request?
An alternative to Openssl would be really helpful for embedded systems.

doublex:
The patch works for me (tested in production). Some issues:

a) Memory leak, this is never free'd:
https://github.com/huitema/picotls/blob/complete-mbedtls-backend/lib/mbedtls_sign.c#L1337

b) Unused variable:
https://github.com/huitema/picotls/blob/complete-mbedtls-backend/lib/mbedtls_sign.c#L406

c) Maybe a typo (`mbedssl` -> `mbedtls`):
`ptls_mbedssl_init_verify_certificate_complete()` -> `ptls_mbedtls_init_verify_certificate_complete()`

doublex:
Patch to fix the memory-leak:
````
*** a/mbedtls_sign.c    2024-09-10 16:56:25.685199324 +0200
--- b/mbedtls_sign.c    2024-09-10 16:56:05.489162611 +0200
*************** static int mbedtls_verify_certificate(pt
*** 1250,1255 ****
--- 1250,1256 ----
  
      if (chain_head != NULL) {
          mbedtls_x509_crt_free(chain_head);
+         free(chain_head);
      }
      return ret;
  }
*************** void ptls_mbedtls_dispose_verify_certifi
*** 1390,1395 ****
--- 1391,1397 ----
      if (verifier != NULL) {
          if (verifier->trust_ca != NULL) {
              mbedtls_x509_crt_free(verifier->trust_ca);
+             free(verifier->trust_ca);
              verifier->trust_ca = NULL;
          }
          if (verifier->trust_crl != NULL) {
````

doublex:
Maybe session-tickets are missing (server): 
````
picotls_encrypt_ticket.cb = encrypt_ticket_cb;
picotls_context.encrypt_ticket = &picotls_encrypt_ticket;
````

doublex:
Update:
Client works (nice, because `minicrypto` fails to verify)

Server fails:
a) Sending stateless-retry-token results in stack-buffer-overflow, this line:
   https://github.com/h2o/picotls/blob/bad0e5077fc4d6673092a4fffec0a5620f6140ad/lib/mbedtls.c#L346
b) After increasing buffer, `curl --http3-only` reports "bad signature"

doublex:
@huitema 
Is the indentation correct?
https://github.com/huitema/picotls/blob/complete-mbedtls-backend/lib/mbedtls_sign.c#L1117

That doesn't look right:
````c
psa_verify_message(..., alg=PSA_ALG_NONE, data.base=NULL, ....);
````

huitema:
I am sorry. I should really close and abandon this PR. Or someone else could take it over.

Since it was never checked in, I developed the corresponding code inside picoquic, see `https://github.com/private-octopus/picoquic/blob/master/picoquic/picoquic_mbedtls.c`. I do not actually need picotls to support MbedTLS, I do not use this port, I just build a mapping library inside the picoquic code.

If someone wants to update the picotls code based on the port shipping in picoquic, they are welcome to do so, and I promise to help them.

PR #536, opened 2024-08-07
Allow LibreSSL to use X25519

nak3:
As LibreSSL already supports X25519 ([changelog](https://github.com/libressl/portable/blob/master/ChangeLog)). Also NID_X25519 variable is also [defined](https://github.com/libressl/openbsd/blob/9d394c7780687b920ba0b1d0c64ec49d7a5a1772/src/lib/libcrypto/evp/evp.h#L120).
Hence, this patch allows LibreSSL to use X25519.


nak3:
Sorry, I misunderstood that it could be available. I will come back after adding the compatible function on LibreSSL if possible.

PR #560, opened 2025-01-21
minicrypto API

doublex:
API proposal for `minicrypto.h`
Works with any combination of `minicrypto`, `openssl`, `fusion` and `libaegis` at compile time without cluttering the code with tons of `#ifdef`

Does not break current code but adds a  thin layer of `#defines`.

Proposed API - with counterparts for `openssl` and `minicrypto`:

`ptls_crypto_init_cipher_suites()`
`ptls_crypto_init_key_exchanges()`

`ptls_crypto_init_sign_file()`
`ptls_crypto_init_sign_der()`
`ptls_crypto_dispose_sign()`

`ptls_crypto_init_verify_file()`
`ptls_crypto_init_verify_der()`
`ptls_crypto_dispose_verify()`


doublex:
The result of these pull requests are 4 files: `minicrypto.h` `minicrypto.c` `sign.c` `verify.c`
They may be easier to read as code files than a bunch of diff's.
[minicrypto.zip](https://github.com/user-attachments/files/18496428/minicrypto.zip)

PR #589, opened 2026-04-08
Fix out-of-bounds read in ASN.1 type-and-length parser

jedisct1:
`ptls_asn1_get_expected_type_and_length()`, which is part of the public API, could read past the end of the input buffer.


kazuho:
Thank you for the fix!

PR #591, opened 2026-04-16
Fix error propagation in TLS 1.2 receive path

jedisct1:
handle_input_tls12 unconditionally returned 0, silently discarding AEAD decryption failures, bad MAC errors, and other fatal conditions on imported/resumed TLS 1.2 connections.

Callers of ptls_receive had no way to observe that a record failed authentication.


kazuho:
Thank you for the fix!

PR #592, opened 2026-04-16
mbedtls: add bounds checks in RSA key-bit parser

jedisct1:
ptls_mbedtls_rsa_get_key_bits accessed key_value[x] through key_value[x+3] without verifying that these indices are within the buffer.

On truncated or malformed DER input this caused out-of-bounds reads before the function could reject the key.


kazuho:
Thank you!

PR #593, opened 2026-04-16
Reject undersized ciphertext in bcrypt AEAD decrypt

jedisct1:
ptls_bcrypt_aead_do_decrypt subtracted the tag size from inlen without first checking that inlen is large enough.

When inlen is smaller than the tag, the size_t subtraction wraps, producing a bogus textLen and an out-of-bounds tag pointer that get forwarded to BCryptDecrypt.


kazuho:
Thank you for the fix!

PR #596, opened 2026-04-17
fix three client-side ECH bugs (grease 0-RTT, HRR/SH, NULL configs)

### LINKED ISSUE #576: Greasing ECH seems to prevent 0RTT
I am testing the following scenario:

- server is not configured to use ECH
- client is configured by setting:
~~~
    ctx->ech.client.ciphers = picoquic_hpke_cipher_suites;
    ctx->ech.client.kems = picoquic_hpke_kems;
~~~

The connection proceeds, with ECH "grease" parameters in the ClientHello. The client obtains a session resume ticket, and starts a new connection with the same parameter and the ticket. The connection succeeds, but 0RTT is not negotiated -- the handshake is not of type PSK.

When I trace what is happening, I see that the session resume ticket (PSK identifier) is copied in the "inner" client hello, and some place holder is copied in the "outer" hello. Of course, since the server is not programmed to do ECH, it cannot accede the inner CH, so no 0RTT.

0RTT does work if the server is programmed to do ECH and if the second connection uses the same ECH parameters. I would have expected that 0RTT will work with ECH Grease, but it does not. That's unfortunate.

kazuho:
## Summary
Three client-side ECH bugs, each with a regression test:

1. GREASE ECH prevented 0-RTT session resumption against non-ECH servers.
2. Per RFC 9849 §6.1.5, a HelloRetryRequest that confirmed ECH followed by a
   ServerHello that does not confirm must abort with `illegal_parameter`;
   picotls silently fell back to the outer ClientHello.
3. `handshake_properties.client.ech.configs = {NULL, 0}` should disable ECH
   (as `picotls.h` documents), but the client was still sending GREASE.

## Details
- **grease 0-RTT, closes #576**: GREASE reused the real-ECH inner/outer
  PSK path, randomizing the outer PSK identity/binder and breaking resumption.
  Refactored to send a normal resumption-capable ClientHello with a dummy but
  syntactically valid ECH extension, matching RFC 9849 §6.2.1.
- **HRR/SH consistency**: `client_ech_select_hello` demoted the post-HRR
  `ACCEPTED` state back to `OFFERED` on an SH confirmation mismatch. It now
  aborts with `PTLS_ALERT_ILLEGAL_PARAMETER`.
- **NULL configs**: the grease branch triggered on `configs.len == 0`
  regardless of `.base`. Added the missing `configs.base != NULL` gate so the
  default zero-initialized `handshake_properties` does not force GREASE.

Plus internal cleanup: the three ECH bit fields collapsed into a single state
enum, test-harness config handling simplified, and binder computation
deduplicated.

## Validation
`build/default/test-openssl.t` — all 17 subtests pass. Each fix has a
dedicated regression test that fails without its corresponding code change.

Closes #576.

### DIFF
--- a/include/picotls.h
+++ b/include/picotls.h
@@ -1123,8 +1123,9 @@ typedef struct st_ptls_handshake_properties_t {
              */
             struct {
                 /**
-                 * Config offered by server e.g., by HTTPS RR. If config.base is non-NULL but config.len is zero, a grease ECH will
-                 * be sent, assuming that X25519-SHA256 KEM and SHA256-AES-128-GCM HPKE cipher is available.
+                 * An ECH config offered by server e.g., by HTTPS RR. If config.len is zero and .base is non-NULL, a grease ECH will
+                 * be sent, assuming that X25519-SHA256 KEM and SHA256-AES-128-GCM HPKE cipher is available. If .base is also NULL,
+                 * ECH will not be used at all, even if the context provided the ECH ciphers.
                  */
                 ptls_iovec_t configs;
                 /**
--- a/lib/picotls.c
+++ b/lib/picotls.c
@@ -1177,7 +1177,6 @@ static void client_setup_ech_grease(struct st_ptls_ech_t *ech, void (*random_byt
                                     ptls_hpke_cipher_suite_t **ciphers, const char *sni_name)
 {
     static const size_t x25519_key_size = 32;
-    uint8_t random_secret[PTLS_AES128_KEY_SIZE + PTLS_AES_IV_SIZE];
 
     /* pick up X25519, AES-128-GCM or bail out */
     for (size_t i = 0; kems[i] != NULL; ++i) {
@@ -1194,10 +1193,6 @@ static void client_setup_ech_grease(struct st_ptls_ech_t *ech, void (*random_byt
     }
     if (ech->kem == NULL || ech->cipher == NULL)
         goto Fail;
-
-    /* aead is generated from random */
-    random_bytes(random_secret, sizeof(random_secret));
-    ech->aead = ptls_aead_new_direct(ech->cipher->aead, 1, random_secret, random_secret + PTLS_AES128_KEY_SIZE);
 
     /* `enc` is random bytes */
     if ((ech->client.enc.base = malloc(x25519_key_size)) == NULL)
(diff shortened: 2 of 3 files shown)


t/picotls.c
```diff
@@ -840,6 +840,14 @@ static int save_client_hello(ptls_on_client_hello_t *self, ptls_t *tls, ptls_on_
     return 0;
 }
 
+enum test_ech_mode {
+    TEST_ECH_NONE,
+    TEST_ECH_REAL,
+    TEST_ECH_GREASE
+};
+
+static enum test_ech_mode test_client_ech_mode = TEST_ECH_REAL;
```
kazuho: Refactoring the `can_ech` function to return a ternary is good, but then, I'm not sure if we need a global variable as well.

t/picotls.c
```diff
@@ -1411,6 +1429,24 @@ static void test_resumption(int different_preferred_key_share, int require_clien
...
+
+    if (get_test_ech_mode(ctx, 0) == TEST_ECH_NONE)
+        return;
+
+    ctx_peer->ech.server.create_opener = NULL;
+    test_client_ech_mode = TEST_ECH_GREASE;
+
+    subtest("basic", test_resumption_impl, 0, 0, 0, 0);
+    subtest("transfer-session", test_resumption_impl, 0, 0, 0, 1);
+    subtest("ticket-request", test_resumption_impl, 0, 0, 1, 0);
```
kazuho: I believe these lines are duplicate code.

lib/picotls.c
```diff
@@ -2341,6 +2336,68 @@ static int encode_client_hello(ptls_context_t *ctx, ptls_buffer_t *sendbuf, enum
...
+
+    *dst = ptls_iovec_init(extbuf.base, extbuf.off);
+    extbuf = (ptls_buffer_t){0};
+    ret = 0;
+
+Exit:
+    ptls_buffer_dispose(&chbuf);
+    ptls_buffer_dispose(&extbuf);
+    return ret;
+}
```
kazuho, line 2371: I'm uncertain if we need to have a function as large as this, just to change where the PSK is stored?

kazuho:
@huitema WDYT?

huitema:
I need to build picoquic against the PR and redo my ECH tests to validate this. Give me a bit of time, please...

lib/picotls.c
```diff
@@ -2375,6 +2397,7 @@ static int send_client_hello(ptls_t *tls, ptls_message_emitter_t *emitter, ptls_
                 /* zero-length config indicates ECH greasing */
                 client_setup_ech_grease(&tls->ech, tls->ctx->random_bytes, tls->ctx->ech.client.kems, tls->ctx->ech.client.ciphers,
                                         sni_name);
+                tls->ech.offered_grease = 1;
```
huitema: If I understand correctly, this code will:

1. try to use ECH if the handshake properties are present and the ech.configs.len is > 0
2. set ECH grease if the handshake properties are present and the ech.configs.len is 0
3. not include the ECH parameter if the handshake properties are not present.

How do we allow the application to *NOT* use ECH and still set handshake properties like the "additional extension" used in QUIC? I am concerned that this code means "using ECH or greasing for QUIC all the time". Is that a deliberate design choice?

lib/picotls.c
huitema: ```suggestion```
         }
         else if (properties->client.ech.configs.base != NULL) {
```
This would allow three states: 
- if ech.configs is set with a non zero string, do ECH
- if ech.configs is set explicitly with a zero length string but a valid base, do grease,
- if ech.configs is left non configured, with NULL base, do not grease.

An alternative would be to add a "do grease" lag in the properties structure.

File: lib/picotls.c (hunk 4 above)
Comment (huitema) [reply lib/picotls.c]: See suggestion above.

**Review by huitema — COMMENTED** [review]
I am not sure of what is happening. The comment in line 844 of t/picotls.c says that "`base == NULL` means no ECH, `len == 0` means grease, `len > 0` means real ECH", but the code in "sendClientHello" does not test for "base == NULL" on line 2403 of picotls.c. Did we miss someting, or am I reading it wrong?

File: include/picotls.h [hunk 5]
```diff
@@ -1123,8 +1123,9 @@ typedef struct st_ptls_handshake_properties_t {
              */
             struct {
                 /**
-                 * Config offered by server e.g., by HTTPS RR. If config.base is non-NULL but config.len is zero, a grease ECH will
-                 * be sent, assuming that X25519-SHA256 KEM and SHA256-AES-128-GCM HPKE cipher is available.
+                 * An ECH config offered by server e.g., by HTTPS RR. If config.len is zero and .base is non-NULL, a grease ECH will
+                 * be sent, assuming that X25519-SHA256 KEM and SHA256-AES-128-GCM HPKE cipher is available. If .base is also NULL,
+                 * ECH will not be used at all, even if the context provided the ECH ciphers.
```
kazuho, line 1128: @huitema Thank you for pointing that out; that has always been the intent according to the header file, but the code has behaved differently.

I've fix the issue, added tests, and have made this doc-comment in the header file explicitly clear about the behavior.

lib/picotls.c
huitema: I see this is fixed now! Thanks.

huitema (approved):
Looks good now! Thanks for fixing this bug.

PR #597, opened 2026-04-17
mbedtls: ED25519 algorithm mapping rejects imported keys

jedisct1:
Imported PKCS8 Ed25519 private keys are assigned `PSA_ALG_PURE_EDDSA` during load, but scheme selection only recognizes `PSA_ALG_ED25519PH` for Ed25519.

As a result, Ed25519 keys are rejected instead of being exposed as usable signing certificates.


kazuho:
Thank you!

PR #598, opened 2026-04-17
mbedtls: RSA PKCS8 import slices wrong DER object

jedisct1:
`ptls_mbedtls_load_private_key` unwraps PKCS#8 `BEGIN PRIVATE KEY` material with `test_parse_private_key_field`, yielding `key_index` and `key_length` for the inner private-key OCTET STRING payload.

In the RSA branch, the code then switches back to the outer PKCS#8 buffer and length before deriving RSA attributes and calling `psa_import_key`.

As a result, RSA-specific parsing receives a `PrivateKeyInfo` object instead of the required inner `RSAPrivateKey` DER object.


kazuho:
Thank you for the PR!

PR #599, opened 2026-04-17
mbedtls: AES-256-CTR misadvertises as AES128-CTR

jedisct1:
`ptls_mbedtls_aes256ctr` is the AES-256-CTR algorithm descriptor, but its public `name` field is set to `"AES128-CTR"`.


kazuho:
Thank you for the fix!

PR #600, opened 2026-04-21
Increase max number of signature algorithms to address Erland RC failure

robguima:
This addresses
[OTP 29.0-rc3: Can't reach erlang.org, fastly.com](https://github.com/erlang/otp/issues/11018)
by increasing the size of the signature array.

The RC build above is currently sending 34 signature algorithms.

CC @ha0li who was involved in the investigation.

I'm not sure the CI test adds much value, but it's there if we want it.


IngelaAndin:
TLS 1.3 RFC defines signature_algorithms as 

struct {
          SignatureScheme supported_signature_algorithms<2..2^16-2>;
      } SignatureSchemeList;

And the RFC itself includes 16 base signature algorithms. And although I think it would be ridiculous having thousands of algorithms it is quite possible you want to have  more than those 16 , for  instance PQC or other examples such as Brainpool signature algorithms.

For the record our project is called Erlang not  Erland  ;)

PR #601, opened 2026-05-07
Check for memory allocation failure in AES-GCM

loganaden:


kazuho:
Thank you very much for the PR. This would be an improvement.

However, the fusion backend has historically assumed allocations do not fail, and I do not think we'd want to change that contract. In particular, `aead_do_encrypt`, which is called by `ptls_aead_encrypt`, may call `ptls_fusion_aesgcm_set_capacity`, which can allocate memory. At that point, we do not have a way to report an error through the AEAD API.

I consider that is the contract between fusion and its users: fusion must only be used when the application can guarantee that sufficient memory is available. In other words, I think we should address the issue by:
* clearly documenting the contract, and
* failing more explicitly, e.g., by calling `abort` if allocation fails.

kazuho:
I've merged the PR as #602, changing abort to assert following the precedence in lib/fusion.c, and adding doc-comments to clarify the contract. Thank you very much @loganaden!

